PR #1, opened 2019-07-05
Completed Day 4 - Please Review

bharatreddypothireddy:
Used recursive binary search approach to find the minimum element in a sorted rotated array.

### DIFF
--- a/MinimumRSA.java
+++ b/MinimumRSA.java
@@ -1,0 +1,37 @@
+public class MinimumRSA {
+    public int findMin(int[] nums) {
+        return search(nums,0,nums.length-1);
+    }
+    public int search(int[] nums,int start,int end){
+        if(start>end)
+            return -1;
+        int mid = start+(end-start)/2;
+        if(nums[mid]<=nums[start]&& nums[mid]<=nums[end]){
+            if(end-start<=2)
+                return nums[mid];
+            else{
+                start = start+1;
+                end = end - 1;
+                return search(nums,start,end);
+            }
+        }
+
+        if(nums[mid]>=nums[start] ){
+            if(nums[mid]>=nums[end])
+                start = mid+1;
+            else
+                end = mid-1;
+        }else{
+            if(nums[mid]<=nums[end])
+                end = mid-1;
+            else
+                start = mid+1;
+        }
+        return search(nums,start,end);
+    }
+    public static void main(String[] args) {
+        MinimumRSA obj = new MinimumRSA();
+        int[] x = new int[]{4,5,1,2,3};
+        System.out.println(obj.findMin(x));
+    }
+}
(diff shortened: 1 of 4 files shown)


MinimumRSA.java
```diff
@@ -0,0 +1,37 @@
...
+            else
+                start = mid+1;
+        }
+        return search(nums,start,end);
+    }
+    public static void main(String[] args) {
+        MinimumRSA obj = new MinimumRSA();
+        int[] x = new int[]{4,5,1,2,3};
+        System.out.println(obj.findMin(x));
+    }
```
super30admin: Check it for the test case [1,2].

MinimumRSA.java
```diff
@@ -0,0 +1,37 @@
...
+        if(start>end)
+            return -1;
+        int mid = start+(end-start)/2;
+        if(nums[mid]<=nums[start]&& nums[mid]<=nums[end]){
+            if(end-start<=2)
+                return nums[mid];
+            else{
+                start = start+1;
+                end = end - 1;
+                return search(nums,start,end);
```
super30admin: Can you do this iteratively? Which solution is better recursive or iterative?

searchRange.java
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    public static void main(String[] args) {
+        searchRange obj = new searchRange();
+        int[] x = new int[]{5,7,7,8,8,10};
+        //System.out.println(x);
+        int[] y = obj.searchRange(x,8);
+        System.out.println(y[0]+"  "+y[1]);
+    }
+}
```
super30admin: Thats a perfect solution.
Cheers!!

PeakElement.java
```diff
@@ -0,0 +1,24 @@
...
+        if(start>end)
+            return -1;
+        int mid = start+(end-start)/2;
+        if((mid==0 || nums[mid]>nums[mid-1]) && (mid==nums.length-1 || nums[mid]>nums[mid+1]))
+            return mid;
+        if(mid>0 && nums[mid]<nums[mid-1])
+            return binarySearch(nums,start,mid-1);
+        else
+            return binarySearch(nums,mid+1,end);
+    }
```
super30admin: Thats a good Solution.
Well Done!!

MinimumRSA.java
bharatreddypothireddy: Works fine for the above test case

MinimumRSA.java
bharatreddypothireddy: Implemented using iterative approach (MinimumRSA2.java). 
In my opinion, both have the same time complexity. But if memory is limited iterative approach is better because for large arrays, using a recursive approach takes more space and may cause StackOverflow .

jaspindersingh83: Thats the great explain for iterative vs recursive approach Bharath

PR #10, opened 2019-07-06
problem 1 and 2

sarang4488:


jaspindersingh83:
Problem 1: Works perfectly sarang. Its an even better solution than we discussed in the class. Try recursive solution as well
Problem 2: Good one
Submit problem 3 as well

PR #14, opened 2019-08-02
Completed All Problems. Please review

priyansh22:


Problem2.cpp
```diff
@@ -0,0 +1,56 @@
...
+    int min = INT_MAX;
+    int mid = left + (right - left) / 2;
+    if (left<=right) {
+        if(left==right)         //if only one element in the array. CASE 1.
+        {
+            cout<<left; return left;
+        }
+
+        
+        if(arr[mid]>arr[mid-1] && arr[mid]<arr[mid+1])          //CASE 2: if middle is > left and smaller than right. Search the right part.
```
kaurramanpreet: this is not correct as this way you are always thinking that the mid is always the point where the array is sorted which is not always the case

Problem2.cpp
```diff
@@ -0,0 +1,56 @@
...
+        }
+
+        
+        if(arr[mid]>arr[mid-1] && arr[mid]<arr[mid+1])          //CASE 2: if middle is > left and smaller than right. Search the right part. 
+                                                                // Example Case : [4,6,8,9,1,2]. Middle is 8. Greater than 6, Less than 9. Search Right.
+        {   return binarySearch(arr, mid + 1, right);
+
+        }
+
+        if(arr[mid]>arr[mid-1] && arr[mid]>arr[mid+1])      //CASE 3 : if middle element is bigger than left side and also right side,
```
kaurramanpreet: same goes here

Problem3.cpp
```diff
@@ -0,0 +1,38 @@
...
+
+
+int binarySearch(int arr[], int left, int right)    //Binary search implemented.
+{
+
+    if (left<=right) {
+        int mid = left + (right - left) / 2;
+
+
+        if (arr[mid]>arr[mid+1]&& arr[mid]>arr[mid-1]) {		// Check if Middle is greater than both left and right side.
```
kaurramanpreet: not a correct way to approach this problem take some testcases then you will be able to get the point

Problem1.cpp
```diff
@@ -0,0 +1,50 @@
...
+    if(arr[mid]==target)			//CASE 1 : If the middle element is equal to target.
+    {
+        if(arr[mid]==arr[mid-1])		// CASE 1.1. Example ; [8,8,9] Target 8. Middle = Middle-1 
+        {
+            cout<<"["<<mid-1<<","<<mid<<"]";
+            return mid;
+        }
+        if(arr[mid]==arr[mid+1])		// CASE 1.2: Example ; [8,9,9] Target 9. Middle = Middle+1
+        {
+            cout<<"["<<mid<<","<<mid+1<<"]"; return mid;
```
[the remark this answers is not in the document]
priyansh22: Oh yes. Understood the error. Will update it. Thank you

PR #15, opened 2019-08-03
Checking MinElement

arnav1996:


findMin.java
```diff
@@ -0,0 +1,31 @@
+class Solution {
+    public int findMin(int[] nums) {
+        int low = 0;
+        int high = nums.length -1;
+
+        while(low<high){
+        int mid = low + (high - low)/2;
+        if(low == high){
```
kaurramanpreet: Do you think that you will ever reach to this condition? Inside while loop?

arnav1996: No

arnav1996: Modified it. Thanks.

PR #17, opened 2019-08-03
Completed Binary-Search-2

AnkitaPatil10:
Kindly review and let me know in case any changes/modifications needs to be done.


kaurramanpreet:
dont you think your approach to search for first and last index of an element is of time complexity O(n)?

AnkitaPatil10:
Yes in case if my entire array consists of target element in that case it will be O(n).
I will try to modify it further. Thanks for the comment.

kaurramanpreet:
great!!

PR #18, opened 2019-08-05
Completed Binary Search-2. Please review.

manogna-mujje:


MinOfRotatedSortedArray.java
```diff
@@ -0,0 +1,34 @@
...
+			while(low < high) {
+	            if(nums[low] == nums[high]) {
+					return nums[low];
+				}
+				mid = low+(high-low)/2;
+				if((mid == 0 || nums[mid] < nums[mid-1]) && nums[mid] < nums[mid+1]) {
+					return nums[mid];
+				} else {
+					//TODO Why can't it nums[low] compared with nums[mid] here?
+					if(nums[high] < nums[mid]) {
```
kaurramanpreet: you cant do it that way because we are moving the low and high pointers by checking that whether we are in first sorted half array or in 2nd because if we are in 1st sorted half then we want to move our low pointer to mid+1 else we want to move our high pointer to mid-1
but if you would do the same thing using low instead of high you wont be able to know whether you are in 1st half or in second half and how to proceed furthur
like lets take an example {4,5,6,7,1,2,3}
initially my mid is at 7
i compared it with 4 i found i am in first sorted array so i moved my low pointer to 1
now my mid is at 2 now i will compare my mid with 1 and i would start thinking the same way as before that i am in 1st sorted array and so to move to the right i need to make my low pointer point to mid+1

but if i do the same thing using high this wont happen

manogna-mujje:
Hi Raman, thank you for the detailed explanation. I still have some doubts on this. Can we have a quick call whenever you’re free after 7pm PST today?

PR #21, opened 2019-08-22
Problems 1 and 2 done

pradhan1234:
Please review @kaurramanpreet @jaspindersingh83

### DIFF
--- a/FindMinRotatedSortedArray.java
+++ b/FindMinRotatedSortedArray.java
@@ -1,0 +1,49 @@
+/**
+ * Sorted Array --> Good Candidate for Binary Search
+ * 
+ * Idea: Using Binary Search to find the minimum element in rotated sorted array.
+ * 
+ * Runtime: 0 ms, faster than 100.00% of Java online submissions for Find
+ * Minimum in Rotated Sorted Array. 
+ * Memory Usage: 37.8 MB, less than 100.00% of
+ * Java online submissions for Find Minimum in Rotated Sorted Array.
+ */
+public class FindMinRotatedSortedArray {
+    /**
+     * Find the minimum element in the given sorted rotated array.
+     * 
+     * @param nums given sorted rotated array
+     * @return minimum element
+     */
+    public int findMin(int[] nums) {
+        // base case checks
+        // 1. array contains only single element
+        // 2. array has not been rotated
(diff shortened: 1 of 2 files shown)


pradhan1234:
Added Partial Solution to FindPeakElement

pradhan1234:
cc @kaurramanpreet fixed the bug, please review

FindPeakElement.java
```diff
@@ -9,8 +9,9 @@
  * 
  * Space Complexity: O(1) Time Complexity: O(log n)
  * 
- * Not yet submitted, the following code contains bugs
- * TODO: Fix the Bug
+ * Leetcode Result:
```
kaurramanpreet: mention this as well!!

FindPeakElement.java
```diff
@@ -20,18 +21,21 @@
...
+        if(nums == null || nums.length == 0) {
+            return -1;
+        }
+        if(nums.length == 1) {
+            return 0;
+        }
+        int low = 0, high = nums.length- 1;
         while (low < high) {
             int mid = low + (high - low) / 2;
-            if ((mid > 0 && nums[mid] > nums[mid - 1]) && (mid < nums.length - 1 && nums[mid] > nums[mid + 1])) {
```
kaurramanpreet: how can you return mid whithout checking whether the value at mid = mid+2 or not?

FindPeakElement.java
pradhan1234: Do you want me to add a description about what I fixed?

FindPeakElement.java
pradhan1234: This is based on the template 2 of binary search that Komal explained in the previous session
Where we do not have a condition check in the while loop and return when the loop terminates.

I didnt understand this part `mid = mid+2` can you please explain?

pradhan1234:
I have checked the code and verified with leetcode as well

FindPeakElement.java
pradhan1234: I think value at mid = mid + 2 is the condition that in the problem of finding the missing element in the sorted array here: 
https://github.com/super30admin/Competitive-Coding-1/pull/19

kaurramanpreet:
Great work
No i wanted you to fix the bugs and check again

PR #23, opened 2019-08-23
completed

PATIL1695:


 Find First and Last Position of Element in Sorted Array.py
```diff
@@ -0,0 +1,51 @@
...
+    def searchRange(self, nums, target):
+        start = 0
+        end = len(nums) -1
+        result = [None] * 2
+        result[0] = -1
+        result[1] = -1
+        if nums == None or len(nums) == 0:
+            return result
+        
+        while (start+1  <= end):
```
kaurramanpreet: why did you use start+1 and not start?

 Find First and Last Position of Element in Sorted Array.py
```diff
@@ -0,0 +1,51 @@
...
+            mid = (start+end) // 2
+            if nums[mid] == target:
```
kaurramanpreet: is this the enough condition?

PATIL1695:
This code passed 84/88 test cases and used start also instead of start+1
but the while loop was running infinitely  when using start. May be because
of the way I have declared mid. Could you please help me correct this code.

Warm regards-
Vishwanath Patil

kaurramanpreet:
whenever you declare mid it should be like mid= left+(right-left)/2;

PR #30, opened 2019-10-09
Incomplete - Couldn't understand how to solve

prajaktaray:


peakElem.py
```diff
@@ -0,0 +1,21 @@
+# Couldn't find the solution for when there is no peak Eg : [8,7,5,2,5,7,8]
+# Didn't understand meaning of nums[-1] = nums[n] = -∞
```
sehaj-grover: basically we are saying is that for [8,7,5,2,5,7,8] we are assuming that the array is [0,8,7,5,2,5,7,8,0] that for the egde values the left and right value is lower

minRotatedArray.py
```diff
@@ -0,0 +1,36 @@
+
+# Cannot think of ALL test cases to check the correctness of the algo
+# Time Com : O(lgn) Space Com : O(1)
+# No link for Leetcode so couldn't run it there
+
```
sehaj-grover: https://leetcode.com/problems/find-minimum-in-rotated-sorted-array/

sehaj-grover:
Did you understand the 3 problems or do you still have any doubts?

prajaktaray:
Yes I understood. Thanks!

PR #33, opened 2019-10-10
Please Review

08vishal:


FInd_Peak_Element.py
```diff
@@ -0,0 +1,23 @@
...
+#LeetCode: no, giving runtime error
+#Problem Faced:how to apply binary search
+class Solution(object):
+    def findPeakElement(self, nums):
+        """
+        :type nums: List[int]
+        :rtype: int
+        """
+        #doing a brute force 
+        if len(nums)==0: return 0
```
sehaj-grover: also check for length ==1 because it will always be the peak

FInd_Peak_Element.py
```diff
@@ -0,0 +1,23 @@
...
+        low=0
+        high=len(nums)-1
+        for i in range(1,len(nums)-1):
+            if nums[i]>nums[i-1] and nums[i]>nums[i+1]:
+                return i
+        if nums[low]>nums[low+1]:
+            return 0
+        if nums[high]>nums[high-1]:
+            return len(nums-1)
```
sehaj-grover: can't substract 1 from list 
it should be len(nums)-1

First_last_position.py
```diff
@@ -0,0 +1,34 @@
...
+class Solution(object):
+    def searchRange(self, nums, target):
+        """
+        :type nums: List[int]
+        :type target: int
+        :rtype: List[int]
+        """
+        def BinarySearchRight(nums,target):
+            l,r=0,len(nums)-1
+            while l<=r:
```
sehaj-grover: if l<=r

PR #36, opened 2019-10-11
First and Last index

88sanjay:


sehaj-grover (commented):
very good sanjay
try the other problems as well

PR #62, opened 2019-11-24
Binary Search 2

iyersneha:


Problem_3.cpp
```diff
@@ -0,0 +1,18 @@
+class Solution {
+public:
+    int findPeakElement(vector<int>& nums) {
+        if (nums.size() == 1) return 0;
+        int left = 0;
+        int right = size(nums) - 1;
+
+        while(left <= right){
```
neha79091: because of = sign mid is assigned infinite times to right  and time limit is exceeding

Problem_2.cpp
```diff
@@ -0,0 +1,26 @@
...
+            if (mid == 0 && (nums[mid] < nums[mid - 1])){
+                return nums[mid];
+            }
+            else if ((nums[low] <= nums[mid])  && (nums[mid] > nums[high])){
+                low = mid + 1;
+            }else{
+                high = mid - 1;
+            }
+
+        }
```
neha79091: Try to solve this problem again. Logics are incorrect.

PR #66, opened 2019-12-26
Binary Search 2 - Kindly Review

SiddharthMehtaRajendra:
Binary Search 2 - Kindly Review

### DIFF
--- a/FirstLastPosition.java
+++ b/FirstLastPosition.java
@@ -1,0 +1,45 @@
+// Time Complexity : O(logn)
+// Space Complexity : Not sure
+// Did this code successfully run on Leetcode : No. 21/88 test case pass.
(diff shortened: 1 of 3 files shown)


rajatkumar01 (commented):
HI Siddharth,
It was nice try
Just try to keep the below point and try to solve it

We can do separate binary search for first and last elements. Why?
Because for finding first occurence we keep doing binary search in left half when we encounter mid that is same as mid-1 element.
We should not iterate linearly towards left and right after finding mid that is equal to target. If we do a linear search then asymptotic time complexity will again become O(n). 
Time Complexity: log(n) Space: O(1)

rajatkumar01 (commented):
Nice Job!!!

rajatkumar01:
Nice Job!!!

PR #67, opened 2019-12-27
Binary-Search-2, Problem1 and 2 completed - Please review

preranas20:


Problem_1.java
```diff
@@ -0,0 +1,68 @@
...
+                    return new int[]{mid,mid};
+            }else if(target<nums[mid])
+                high = mid -1;
+            else 
+                low = mid + 1;
+        }
+        
+        return new int[]{-1,-1};
+    }
+}
```
jaspindersingh83: Too much redudant code here. Try this solution after the class as well

Problem_2.java
```diff
@@ -0,0 +1,53 @@
...
+    public int findMin(int[] nums) {
+        
+        //if array has just one value
+        if(nums.length == 1)
+            return nums[0];
+        
+        //if array has just 2 values
+        if(nums.length == 2)
+            return (nums[0]<nums[1])?nums[0]:nums[1];
+
```
jaspindersingh83: No need of handling array of size 2 as edge case handle it in the main logic

Problem_2.java
```diff
@@ -0,0 +1,53 @@
...
+       
+        
+        while(low<=high){
+            
+            
+            int mid = (high-low)/2+low;
+            
+            if(nums[mid] > nums[mid+1])
+               return nums[mid+1];
+
```
jaspindersingh83: It doesn't go out of bounds on Leetcode?

PR #69, opened 2019-12-27
Please review only problem 153 and 162. 

nikrasiya:


162_find_peak_element.py
```diff
@@ -0,0 +1,32 @@
...
+            (if it is not greater than its neighbour) which ever is greater than it
+            we go towards that because doing so we are 100% sure there are only two
+            options ie. a number can be greater than it or lesser than it. We keep
+            doing this recursively.
+        """
+        start = 0
+        end = len(nums) - 1
+        while start < end:
+            mid = (start + end) // 2
+            if nums[mid - 1] < nums[mid] and nums[mid + 1] < nums[mid]:
```
jaspindersingh83: You are not getting any array out of index exception here?

162_find_peak_element.py
```diff
@@ -0,0 +1,32 @@
...
+                return mid
```
jaspindersingh83: No array out of index exception here?

162_find_peak_element.py
nikrasiya: the strange way how python works. When we are at index 0 and we reduce 1 it goes to -1, which is a valid index in python. But I think I should still make it clear, will not use any language-specific feature as instructed by you. I will update it in some time.

super30admin:
Good Job

PR #74, opened 2019-12-27
Min element in a Rotated sorted Array, First and Last Position in a sorted array and Peak element in a array using Binary Search( Please review)

ShreyasP7495:


ques9.cpp
```diff
@@ -0,0 +1,40 @@
+Time Complexity-O(log n)
+Space Complexity-O(n)
+Did the code execute successfully on Leetcode-Yes
```
jaspindersingh83: The space complexity is O 1

ShreyasP7495: I'm sorry i will update that.
It was by mistake.

Thanks for the update.

PR #94, opened 2020-02-06
Completed first 2 questions in logarithmic time -- Please Review

AbhiniveshP:
Hi,

I've completed the first two questions in logn time and constant space. 

Thanks,
Abhinivesh

### DIFF
--- a/First_Last_Range.py
+++ b/First_Last_Range.py
@@ -1,0 +1,55 @@
+class Solution(object):
+    def searchRange(self, nums, target):
+        """
+        Solution:
+        1. First find whether element is present of not. If not present return [-1, -1]
+        2. If present, consider the index found(mid) to be the range [mid, mid] and store the chunk where you found mid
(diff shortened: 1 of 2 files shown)


AbhiniveshP:
Hi,

I have completed find peak elements in logarithmic time as well along with the first two questions. But, took a lot of effort to perform all edge cases correctly. If any of the questions is not optimal enough if you think, feel free to comment on the same.

Also, encountered one small issue with Python while doing [mid-1] index when mid is 0, mid-1 would become -1 and if we try to access arr[mid-1] i.e. arr[-1], then that would not throw an error of Index out of Bounds Exception instead it accesses the last element of the array. You can remind the Python students about this in the class tomorrow as it might turn out to be a problem just the way it happened to me.

super30admin:
Good Job with the PR Abhivinesh. I hope your doubts regarding edge case, time and space are addressed after the class

AbhiniveshP:
Yes, it is clear after the class. Thank You !!!

PR #129, opened 2020-03-17
Binary search 2 completed

vaidyapawan:
153. Find Minimum in rotated sorted array
162. Find a peak element
34. Find first and last position of an element in sorted array


devashishnyati:
Good job Pawan!

PR #141, opened 2020-04-24
Submitted PR-Please review

sbsreedh:


findMin.py
```diff
@@ -0,0 +1,73 @@
+
+#Time Complexity=O(log n)
+#Space Complexity=O(1)
+# Diffciulties faced: I am using min() here , does that affect the time complexity?
```
iivaishnavii, line 4: Using built in methods do add up to complexity. 
In this case, using min is not advisable as you can just return min(nums) without even doing the binary search bit.

searchRange.py
```diff
@@ -0,0 +1,46 @@
...
+        :type nums: List[int]
+        :type target: int
+        :rtype: List[int]
+        """
+        
+        def firstIndex(nums, target):
+            start, end = 0, len(nums) - 1
+            while start <= end:
+                mid = start + (end - start) // 2
+                if nums[mid] == target:
```
iivaishnavii, line 24: Interesting way of handling limits.

iivaishnavii:
Great job,Sonali!

PR #144, opened 2020-04-24
PR for Aditya D

aditya-di:
Includes examples 7 and 8.

### DIFF
--- a/7FindFirstandLastElement.py
+++ b/7FindFirstandLastElement.py
@@ -1,0 +1,58 @@
+"""
+LC:34
+#T:O(n)
+"""
+
+#Brute force
+#T:O(n)
+"""
+def getIndex(nums,target):
+  res=[]
+  arr=nums
+  for i in range(len(nums)):
+    j=i+1
+    if arr[i]==target and j<len(arr) and arr[j]==target :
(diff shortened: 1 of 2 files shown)


7FindFirstandLastElement.py
```diff
@@ -0,0 +1,58 @@
...
+def getIndex(nums,target,l,h):
+  mid=int(l+h/2)
+  result=[]
+  #base
+  while(l<=h):
+   print(nums[l],nums[h],nums[mid])
+   if target==nums[mid]:
+     result.append(mid)
+     previous=mid-1
+     nxt=mid+1
```
super30admin, line 31: What if the right most index is 5 elements away from current mid.

super30admin:
Solution of first and last element search should not be working on LC. Can you confirm.

Please submit logn solution for find min

7FindFirstandLastElement.py
aditya-di: Will make changes in the code to consider [3,3,3]

aditya-di: Made changes in Problem 7-Made first and last index.

PR #151, opened 2020-04-25
Binary-Search-2, please review

Siyami:


super30admin:
Good Job Siyami.

Try to modularize your code in search range function. Your code works but is less readable

PR #163, opened 2020-05-06
Completed Binary-Seach-2 : Please Review

SoumyaSreedhar:


Find Peak Element.py
```diff
@@ -0,0 +1,25 @@
+#162. Find Peak Element
+#Time Complexity -> O(Log n)
```
iivaishnavii, line 2: Define what n is too.

iivaishnavii:
Good job, Soumya!

PR #170, opened 2020-05-27
COmpleted 2 problems

GahanJagadeesh:


findPeak.java
```diff
@@ -0,0 +1,22 @@
+class Solution {
```
Lohith-Aswathappa, line 1: Add time and space complexities.

Lohith-Aswathappa:
Well done, Gahan!

PR #186, opened 2020-05-28
added all files

Akash-Akki:
Solved

1. Find the first and last position of a element in a sorted array.
2. Find minimum in a rotated sorted array
3. Find Peak Element


chiragshahi:
Good job!
Kindly follow the start template for the problems

Akash-Akki:
Yea will do that.
Since I joined two days late for the I could not get the email of first two
days. Can you send the email of the first two days as well

Regards,
Akash Akki

PR #229, opened 2020-07-10
Completed all 3 problems. Kindly Review it !

sneharnair:


Problem-3_Peak_element.py
```diff
@@ -0,0 +1,31 @@
...
+            return -1
+        
+        start , end = 0, len(nums) - 1
+        while start <= end:
+            mid = start + (end - start) // 2
+            
+            if (mid == 0 or nums[mid] > nums[mid - 1]) and (mid == len(nums) - 1 or nums[mid] > nums[mid + 1]):
+                return mid
+            
+            elif nums[mid] < nums[mid + 1]:
```
samarthjj, line 25: Please perform bounds check, what if mid here is len-1.

Problem-2_Minimum_in_rotated_sorted_array.py
```diff
@@ -0,0 +1,35 @@
...
+            return nums[0]
+        
+        start, end = 0, len(nums) - 1
+        while start <= end:
+            mid = start + (end - start) // 2
+            
+            if (mid == 0 or nums[mid] < nums[mid - 1]) and (mid == len(nums) - 1 or nums[mid] < nums[mid + 1]):
+                return nums[mid]
+            
+            elif nums[mid] < nums[0]:
```
samarthjj, line 28: This should be **nums[mid] < nums[end]**
This will not work if for the first iteration start is moved to mid +1, removing index 0 from the search space.

samarthjj:
Nice work Sneha, please review the inline comments.

samarthjj:
Nice work Sneha, please review the inline comments.

Problem-3_Peak_element.py
sneharnair: I have checked if mid is len(nums) - 1 in second part of and condition !

samarthjj: Yes, but that has an **or** condition with it, which doesn't mean that **mid == len(nums)-1** will always be satisfied!
Please reply to inline comments to slack.

PR #256, opened 2020-07-18
completed. please review.

satishsolanki1990:


samarthjj:
Good work Satish, Please also include your approach in 3-4 bullet points for each question!

samarthjj:
Good work Satish, Please also include your approach in 3-4 bullet points for each question!

PR #323, opened 2020-09-18
Binary-Search-2 solution 1,2 working 3 failing on edge cases

krunalT21:


Solution1.java
```diff
@@ -0,0 +1,55 @@
+class Solution {
+    public int[] searchRange(int[] nums, int target) {
```
kannan-mrugapphan, line 2: Basically the logic for firstOccurence() and lastOccurence() is same. I would recommend you to use a flag and avoid duplication of code.

kannan-mrugapphan (commented):
Good Job. Reach out to me if 3rd problem is still unresolved.

kannan-mrugapphan:
Good Job

Solution1.java
krunalT21: okay

PR #328, opened 2020-09-18
Binary Search-2 submission

srinidhib-cmd:
Hi, 
Please check my submission. 
For one of the questions, I was able to come up with only brute force - Will update that after class again. 

- Thanks

### DIFF
--- a/first_last_position_lc34.py
+++ b/first_last_position_lc34.py
@@ -1,0 +1,31 @@
+"""
+Author: Srinidhi
+Did the problem run on leetcode: No - I got TLE as I used brute force
+
+Time Complexity: 
+Space Complexity:
+
+Logic: Find the first and last position using linear scan and mark the first and last positions in the result array
+"""
+
+class Solution:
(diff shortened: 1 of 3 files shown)


first_last_position_lc34.py
```diff
@@ -0,0 +1,31 @@
+"""
+Author: Srinidhi
+Did the problem run on leetcode: No - I got TLE as I used brute force
+
+Time Complexity: 
+Space Complexity:
+
+Logic: Find the first and last position using linear scan and mark the first and last positions in the result array
+"""
+
```
Lohith-Aswathappa, line 10: solve using the approach taught in class

Lohith-Aswathappa (commented):
Nice try, Srinidhi!

Lohith-Aswathappa:
Nice try, Srinidhi!

first_last_position_lc34.py
srinidhib-cmd: Yes Lohit, I will be updating this question soon.

PR #330, opened 2020-09-18
completed 2 questions - pLease review

Vjais:


firstandlast.java
```diff
@@ -0,0 +1,44 @@
+O(n),O(1)
+public int search(int[] nums, int target) {
```
kannan-mrugapphan, line 2: You are just doing binary search. So time is O(log n).
The logic for firstOccurence() and lastOccurence() is same. I would recommend you to use a flag and avoid code duplication.

kannan-mrugapphan (commented):
Good Job. Please submit the other problem as well.

kannan-mrugapphan:
Good Job. Please submit the other problem as well.

PR #332, opened 2020-09-18
Ques 1 and 2 attempted

meetdave06:


meetdave06:
Attempted the questions again

PR #339, opened 2020-09-18
Completed Sprint Binary_Search_2 - Please review

arzoouzma01:


Problem-1.java
```diff
@@ -0,0 +1,50 @@
...
+/*
+Any problem you faced while coding this : None
+
+*/
+
+// Your code here along with comments explaining your approach
+//two pointer approach where you search the target from the front and back
+class Solution {
+    //find the first occurance of the target
+    public int firstOccurance(int[] nums, int low, int high, int target)
```
kannan-mrugapphan, line 17: I would recommend you to have a flag and reuse the findFirstOccurence() instead of repeating the same logic in another method

kannan-mrugapphan (commented):
Nice Explanation. Good Job

kannan-mrugapphan:
Nice Explanation. Good Job

PR #347, opened 2020-09-24
Adding Binary Search 2 problems solution

monakhandat:
I will write the optimized code for problem 3. The current one is a placeholder.


kannan-mrugapphan (commented):
Good Job. The space is constant for the first 2 problems.

kannan-mrugapphan:
Good Job. The space is constant for the first 2 problems.

PR #369, opened 2020-10-16
Binary-Search-2 Submission

sumitjanawlekar:


Problem1.java
```diff
@@ -0,0 +1,81 @@
+// Time Complexity :O(log(n))
+// Space Complexity :O(1)
+// Did this code successfully run on Leetcode : yes
```
kannan-mrugapphan, line 3: If you observe, the logic for firstOccurence and lastOccurence are same. Use a flag to avoid code duplication

Problem3.java
```diff
@@ -0,0 +1,40 @@
+// Time Complexity : O(log(n))
+// Space Complexity :O(1)
+// Did this code successfully run on Leetcode : yes
+// Any problem you faced while coding this : Did not understand what does it mean when the question says, nums[-1] == nums[n] == -Infinity ???
```
kannan-mrugapphan, line 4: Its just an assumption. Just consider the -inf is before the first element i.e nums[-1] and after last element i.e nums[n]

kannan-mrugapphan (commented):
Good Job.

kannan-mrugapphan:
Good Job.

PR #377, opened 2020-10-18
Binary Search-1 Completed

nmtraboulsi:


FindFirstandLast.java
```diff
@@ -0,0 +1,45 @@
+class Solution {
+    private int binarySearchLeft(int [] nums, int target) {
```
kannan-mrugapphan, line 2: Please use a binary flag and avoid code duplication.

kannan-mrugapphan (commented):
Good Job. Please include the complexities as well.

kannan-mrugapphan:
Good Job. Please include the complexities as well.

PR #436, opened 2020-12-25
Added Binary Search 2 code

amby2964:


maithrreye:
Good job ! You will understand about edge cases over practice and solving multiple problems.

maithrreye:
Good job ! You will understand about edge cases over practice and solving multiple problems.

PR #449, opened 2020-12-29
Submitting Binary-Search-2 

arjunprakash05:


ramya-sruthi-d (commented):
The code is fine, the sc for all problems is O(1) because you are not using any extra space.

ramya-sruthi-d:
The code is fine, the sc for all problems is O(1) because you are not using any extra space.

PR #451, opened 2020-12-29
Completed Binary Search - 2. Please review

gayathrig269:


kannan-mrugapphan (commented):
Good Job. In findFirstAndLastOccurence, try to use a flag and avoid code duplication caused by lastBinarySearch

kannan-mrugapphan:
Good Job. In findFirstAndLastOccurence, try to use a flag and avoid code duplication caused by lastBinarySearch

PR #588, opened 2021-05-28
Completed Binary-Search-2 module , please review

harshaGitHubDev:


firstAndLastPosition.java
```diff
@@ -0,0 +1,86 @@
...
+      return ans;
+        
+    }
+}
+
+*/
+
+
+class Solution {
+    public int findFirst(int[] nums, int target ){
```
kannan-mrugapphan, line 34: Use a flag to avoid code duplication

kannan-mrugapphan (commented):
Good Job. Nice with the brute force approaches as well.

kannan-mrugapphan:
Good Job. Nice with the brute force approaches as well.

PR #623, opened 2021-05-31
Binary-Search-2 Kindly Review

sridevi0816:


kannan-mrugapphan (commented):
Good Job. Add a flag to avoid code duplication in the first problem.

kannan-mrugapphan:
Good Job. Add a flag to avoid code duplication in the first problem.

PR #652, opened 2021-06-03
Binary Search 2 

hharshitarora:
For these BS problems I had the most difficulty in understanding edge cases.
Specially P3 were mid == 0 and mid == n-1 there specifically to tackle a 2 element array ?

### DIFF
--- a/.idea/.gitignore
+++ b/.idea/.gitignore
@@ -1,0 +1,3 @@
+# Default ignored files
+/shelf/
+/workspace.xml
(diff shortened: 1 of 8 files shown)


Rooppesh:
In P3, the edge case is basically wherein the element lies in either of the ends of the array. Do let me know if you have any specific doubt.

Rooppesh:
In P3, the edge case is basically wherein the element lies in either of the ends of the array. Do let me know if you have any specific doubt.

PR #693, opened 2021-07-23
Binary Search -2 Code for review

rajeevag24:


Rooppesh:
Goo Job Rajeev! Please add the TC and SC as well!

Rooppesh:
Goo Job Rajeev! Please add the TC and SC as well!

PR #695, opened 2021-07-24
Completed Binary-Search-2 (Please Review)

sayantiroy90:


Rooppesh:
Good Job! Do add the TC, SC and Approach used!

Rooppesh:
Good Job! Do add the TC, SC and Approach used!

PR #705, opened 2021-07-25
Binary Search2: Please Review

saman-akhtar:


firstLstPos.py
```diff
@@ -0,0 +1,42 @@
+class Solution:
```
sridhar-anish, line 1: Make sure to include TC and C

sridhar-anish (commented):
Good Job with everything!

sridhar-anish:
Good Job with everything!

PR #720, opened 2021-08-18
Binary-Search-2, Pls review. Thanks

ysrinu1:


Rooppesh:
Good Job! Please add the TC, SC and Approach.

Rooppesh:
Good Job! Please add the TC, SC and Approach.

PR #730, opened 2021-08-20
Binary Search-2 completed, Please review

shilpagowda812:


sridhar-anish:
Good Job with everything, make sure to write TC and SC

PR #780, opened 2021-09-21
chore: BS2

SruthiSathya:
was able to do 
Find Minimum in Rotated Sorted Array in Time complexity O(logN)
Find Peak Element in Timple complexity O(N)
Find First And the Last Position of Element in Sorted Array - did not do.


SruthiSathya:
did not add Find First And the Last Position of Element in Sorted Array

PR #790, opened 2021-09-22
binary search 2 problem 8,9,10.

rushilshah121198:


Rooppesh:
Good Job! Add the TC and SC as well!

Rooppesh:
Good Job! Add the TC and SC as well!

PR #856, opened 2021-12-02
find first and last index 34, min in rotated sorted array 153, find peak of array 162 completed

ajaydevmane:


Stnagam:
Please add comments, TC SC to all the problems!

Stnagam:
Please add comments, TC SC to all the problems!

PR #926, opened 2022-02-03
completed Binary-Search-2 : please review

Parikshit97:


ehassah10 (commented):
Good work! Please include brief explanation about the approach in each solution

ehassah10:
Good work! Please include brief explanation about the approach in each solution

PR #949, opened 2022-02-11
Completed BS-2

IndlamudiSaiTejaswi:


ehassah10 (commented):
Good job!! Please mention briefly the approach taken to solve the problem

PR #1149, opened 2022-06-25
Done Binary Search 2

aneesh2411:


daredevilteja (changes requested):
Please mention TC and SC

daredevilteja:
Please mention TC and SC

PR #1177, opened 2022-06-27
binary serach 2 please review

bpattarkine:


daredevilteja (changes requested):
Please mention TC and SC and also mention if it works in leetcode

daredevilteja:
Please mention TC and SC and also mention if it works in leetcode before raising the PR

PR #1231, opened 2022-07-27
Binary-Search-2 Please review it

vishnuvryeruva:


ishika2741 (commented):
Good work! Kindly include tc and sc as well.

ishika2741:
Good work! Kindly include tc and sc as well.

PR #1233, opened 2022-07-28
completed Binary Search - 2 Please Review

yash-bhargava18:


firstlast.py
```diff
@@ -0,0 +1,53 @@
+#time complexity = O(lon(n))
```
daredevilteja, line 1: TC is O(log n)

minrotated.py
```diff
@@ -0,0 +1,19 @@
+#time complexity = O(lon(n))
```
daredevilteja, line 1: TC is O(log n)

peak.py
```diff
@@ -0,0 +1,20 @@
+#time complexity = O(lon(n))
```
daredevilteja, line 1: TC is O(log n)

daredevilteja:
Good Work! Keep it up

PR #1282, opened 2022-09-10
Solutions for binary search 2

kundankarthik77:


daredevilteja:
Please mention TC, SC and whether it ran successfully on leetcode or not, for every problem

PR #1416, opened 2023-01-27
Completed Binary Search 2 - Please review. 

Design55:


daredevilteja (changes requested):
Since you are using recursion while performing the binary search, the recursion stack would take extra space. So, here the SC would be O(log n) for all the problems.

daredevilteja:
Please do the requested changes.

Design55:
Created the adjustments.

PR #1418, opened 2023-01-28
finished binary2-please review

summerhanyuezheng:


daredevilteja:
Please mention TC, SC and whether it ran successfully in leetcode or not, for every problem.

PR #1473, opened 2023-03-27
Create first&lastpos

Madhura-19:


pvk3081 (commented):
Please add remaining two problems

pvk3081:
Please add remaining two problems

PR #1587, opened 2023-08-21
Binary-Search-2

ghost:
please review

### DIFF
--- a/findPeak.java
+++ b/findPeak.java
@@ -1,0 +1,37 @@
+// Time Complexity : O(log n)
+// Space Complexity : O(1)
+// Did this code successfully run on Leetcode : yes
+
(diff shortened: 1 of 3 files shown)


Nandha2Tech (commented):
Great Work!
In firstLastSortedArray problem, to further optimize, to find the last index,  instead of searching the whole input again, we can pass the first index and the binary search can happen from that index to the last.
In this way TC would be O(log n) + O(log m).

Nandha2Tech:
Great Work!
In firstLastSortedArray problem, to further optimize, to find the last index,  instead of searching the whole input again, we can pass the first index and the binary search can happen from that index to the last.
In this way TC would be O(log n) + O(log m).

PR #1621, opened 2023-09-30
completed binary search-2

sahithsanam:


ishika2741 (commented):
Good work! Kindly add TC and SC for first problem as well.

ishika2741:
Good work! Kindly add TC and SC for first problem as well.

PR #1625, opened 2023-09-30
Completed Binary-Search-2

Mohini-Mayekar:
Completed Binary-Search-2

Problem#9 - First and Last position of element
Problem#10 - Find Min in rotated sorted array
Problem#11 - Find peak element


src/main/java/FindFirstAndLastIndexes.java
```diff
@@ -0,0 +1,84 @@
+/* Time Complexity : O(log (m*n))
```
ishika2741, line 1: check tc once , it would not be log(m*n)

ishika2741 (commented):
Good work!

ishika2741:
Good work!

PR #1635, opened 2023-10-01
Create Find_Peak_Element.py

KrishanuSinha:
A peak element is an element that is strictly greater than its neighbors.

Given a 0-indexed integer array nums, find a peak element, and return its index. If the array contains multiple peaks, return the index to any of the peaks.

You may imagine that nums[-1] = nums[n] = -∞. In other words, an element is always considered to be strictly greater than a neighbor that is outside the array.

You must write an algorithm that runs in O(log n) time.

### DIFF
--- a/Find_Peak_Element.py
+++ b/Find_Peak_Element.py
@@ -1,0 +1,13 @@
+class Solution:
+    def Find_Peak_Element(self,nums:list)->int:
+        l,r = 0, len(nums)-1
+        while l<=r:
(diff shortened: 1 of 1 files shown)


keshav-Gupta-IS (commented):
Good Work, Krishanu!! Keep it up.

keshav-Gupta-IS:
Good Work, Krishanu!! Keep it up.

PR #1815, opened 2024-06-21
binary search 2 homework completed

dhikki-study:
i added sample file to upload code as it was not present

##############34. Find First and Last Position of Element in Sorted Array####################
// Time Complexity : logN
// Space Complexity : N
// Did this code successfully run on Leetcode : following case failed and other may fail
[2,2]
// Any problem you faced while coding this : used binary search and my code worked assuming all element are in increasing order and only 2 elements are present which are duplicate

##############153. Find Minimum in Rotated Sorted Array####################
// Time Complexity : logN
// Space Complexity : N
// Did this code successfully run on Leetcode : Yes
// Any problem you faced while coding this : no problem but 1 of scenario was giving problem where the minimum element was 1 step ahead of mid and to handle it I applied clause mentioned below as #point check

##############162. Find Peak Element#########################
// Time Complexity : N (couldn't find BS way so applied brute force)
// Space Complexity : N
// Did this code successfully run on Leetcode : yes but brute force way
// Any problem you faced while coding this : no

### DIFF
--- a/Sample
+++ b/Sample
@@ -1,0 +1,91 @@
+##############34. Find First and Last Position of Element in Sorted Array####################
+
+// Time Complexity : logN
+// Space Complexity : N
+// Did this code successfully run on Leetcode : following case failed and other may fail
+[2,2]
+// Any problem you faced while coding this : used binary search and my code worked assuming all element are in increasing order and only 2 elements are present which are duplicate
+
+
+// Your code here along with comments explaining your approach in three sentences only
+
+
+class Solution:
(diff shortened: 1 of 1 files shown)


Nandha2Tech (commented):
Please try to check the logic handled in finding the first and last position of an element in sorted array. First and last position is not at all checked.

Nandha2Tech:
Please try to check the logic handled in finding the first and last position of an element in sorted array. First and last position is not at all checked.

PR #1850, opened 2024-07-08
done with binary search 2

BhavyaPranilaSathi:


Nandha2Tech (commented):
Great Work! TC for findfirstandlastposition of elemnet in sorted array would be log(n).

Nandha2Tech:
Great Work! TC for findfirstandlastposition of elemnet in sorted array would be log(n).

PR #1858, opened 2024-07-27
Completed BS2

dharaneesk:


FirstNLastPos.java
```diff
@@ -0,0 +1,53 @@
+//TC: O(Log (n*k))
```
ishika2741, line 1: can you explain your variables , what is k here?

ishika2741 (commented):
Good work!

ishika2741:
Good work!

FirstNLastPos.java
dharaneesk: N would be the length of the input array to find the first element & 
K would be the length of the array from the first element index to the last.

PR #1900, opened 2024-10-02
Completed

Ameya-20:


Nandha2Tech (commented):
Great Work! In problem1, once you identify the target and if you expand on both sides for first and last index, then in the worst case it would become a linear search.

Nandha2Tech:
Great Work! In problem1, once you identify the target and if you expand on both sides for first and last index, then in the worst case it would become a linear search.

PR #1931, opened 2024-12-28
completed binary search 2

shubhakari:


Erlybird:
Good job! PLease submit the 3rd problem too. Please mention the TC and SC at the top of the solution

Erlybird:
Good job! PLease submit the 3rd problem too. Please mention the TC and SC at the top of the solution

PR #1940, opened 2024-12-29
binary search 2 done

triptighanghas:


Erlybird:
In the second problem, min in Rotated Sorted Array. You must return nums[mid] instead of mid.

Erlybird:
In the second problem, min in Rotated Sorted Array. You must return nums[mid] instead of mid.

PR #2078, opened 2025-05-16
Completed Binary Search - 2

priyayeni:


Nandha2Tech (commented):
In the problem FirstAndLastPosition - the logic implemented will behave in O(n) time for worst case scenarios. Hence the time complexity will become O(n). Kindly try to optimize it to O(log n).

Nandha2Tech:
In the problem FirstAndLastPosition - the logic implemented will behave in O(n) time for worst case scenarios. Hence the time complexity will become O(n). Kindly try to optimize it to O(log n).

PR #2128, opened 2025-05-22
Finished binary search 2

utkarsh9:


dalvirushikesh:
Great work! Just fix typos and change res(res, nums[m]) -> min(res, nums[m]).

PR #2155, opened 2025-06-20
Binary-Search 2 done

GuptaDivij:


super30admin:
Strengths:
- Good understanding of binary search and its variations.
- Clear and concise code with appropriate comments.
- Proper handling of edge cases in problem2.java and problem3.java.
- Correct use of time and space complexity analysis.

Areas for Improvement:
- In problem1.java, the findFirst and findLast functions should immediately return when they find the first or last occurrence of the target. Currently, they continue searching even after finding a potential candidate, which can lead to incorrect results in some cases.
- Consider adding more test cases to verify edge cases, especially for problem1.java.
- The variable names could be more descriptive in some places (e.g., 'ans' could be 'result').

GuptaDivij:
Hi,

Thanks for the comments. I wanted to understand how can we directly return
the result for problem 1 if we are finding the first or the last element,
it is not found till the loop ends as once we do find the target, we still
need to keep checking if mid is pointing to the first element, last element
or somewhere in between. Please let me know how I can do it in a better way.

Best,
Divij

PR #2171, opened 2025-06-22
Binary-Search-2 Complete

bhavanamalaga:


super30admin:
Strengths:
1. Correct implementation of binary search to find the first and last occurrences of the target.
2. Good use of a helper function with a boolean flag to differentiate between finding the first and last occurrence.
3. Clear and concise comments explaining the approach.
4. Proper handling of edge cases (e.g., when the target is not found).

Areas for Improvement:
1. The submission includes an unrelated solution (find_min_in_rotated_sorted_array.cpp), which should be separated to avoid confusion.
2. The comments could be slightly more detailed to explain the logic behind the boolean flag in the binary search function.
3. Consider adding a brief explanation of the overall approach in the class or main function comments.

PR #2184, opened 2025-06-25
completed binary search 2

harsh23raval:


super30admin:
- Strengths:
  - The student has correctly implemented the binary search logic for both first and last occurrences.
  - The solution handles edge cases such as an empty array or when the target is not found.
- Areas for Improvement:
  - The method `binarySearchLast` could start the search from the `first` index found (as in the reference solution) to optimize further, though this is a minor improvement.
  - The return value `999999` in `FindMinimumInSortedArray.java` is not ideal; it would be better to return `-1` or throw an exception if no minimum is found, though this is not part of the evaluated problem.
  - The comments in `FindMinimumInSortedArray.java` and `FindPeakElement.java` are not relevant to the evaluated problem (`FirstAndLastOccurance.java`), so they should be omitted if not part of the submission.

PR #2193, opened 2025-07-06
Binary Search 2

botlamanoj29:
Binary Search 2, Solved 3 problems and submitted in leet code.


super30admin:
Strengths:
- The code is well-structured and readable.
- The logic for finding the first and last positions is correct.
- The space complexity is optimal.

Areas for Improvement:
- The time complexity can be improved by avoiding linear search within binary search. Instead, use binary search to directly find the first and last occurrences.
- The solution should focus solely on the problem of finding the first and last positions of a target in a sorted array, as the other files (`FindMinInSortedArray.py` and `FindPeakElement.py`) are not relevant to this problem.
- The edge case handling (e.g., empty array) should return `[-1, -1]` instead of `(-1, -1)` to match the expected output format.

PR #2252, opened 2025-10-04
Binary-Search-2

ashishbadhe97:


super30admin:
- The conditions in `binarySearchFirst` and `binarySearchLast` should be simplified to check if the current element is the first or last occurrence of the target. For example, in `binarySearchFirst`, you should check if `mid == 0` or `nums[mid - 1] != target` to confirm it's the first occurrence. Similarly, in `binarySearchLast`, check if `mid == nums.length - 1` or `nums[mid + 1] != target` to confirm it's the last occurrence.
- The comments are helpful, but the logic for checking the first and last occurrences should be clarified to avoid confusion.
- The return value of `5001` in `findMin` is not necessary and could be replaced with a more meaningful value or handled differently.

PR #2273, opened 2025-11-16
problem 2 and 3

anilkumarm0921-oss:


super30admin:
Strengths:
- The logic for both problems is correct and follows the binary search approach efficiently.
- The time and space complexity are optimal, matching the reference solutions.
- The student has provided clear points explaining the logic, which shows good understanding.

Areas for Improvement:
- Remove arbitrary return values like 77 and 99. Instead, handle edge cases properly or return -1 if no solution is found.
- Improve code formatting and indentation for better readability.
- Add comments to explain the logic within the code, not just as separate points.
- Ensure all edge cases are considered, such as when the array is empty or has only one element.

PR #2302, opened 2026-01-08
Find min, peak and range of an element in rotated sorted array

paridhimalviya:
Problem 1: (https://leetcode.com/problems/find-first-and-last-position-of-element-in-sorted-array/)
Problem 2: (https://leetcode.com/problems/find-minimum-in-rotated-sorted-array/)
Problem 3: (https://leetcode.com/problems/find-peak-element/)


super30admin:
- Strengths:
  - The code is well-commented and follows a logical structure.
  - The approach to solving the problem is correct in principle, using binary search to find the first and last occurrences of the target.
  - The separation of concerns into helper functions (`binarySearchfirst` and `binarySearchLast`) is a good practice.

- Areas for Improvement:
  - The `binarySearchLast` function is incomplete and contains a critical bug. It needs to include the `else if` and `else` conditions to handle cases where the target is not found or is less than/more than the middle element.
  - The initial check for `nums.count == 0` is good, but the check for `target < nums[0] || target > nums[nums.count - 1]` could be moved to the beginning of the `searchRange` function to avoid unnecessary computations.
  - The variable names `lowIndex` and `highIndex` in `searchRange` are unused and should be removed.
  - The solution could benefit from more thorough testing, especially edge cases like empty arrays, single-element arrays, and arrays where the target is not present.

super30admin:
Strengths:
- The code is well-commented and follows a logical structure.
- The `binarySearchfirst` function is correctly implemented and handles edge cases well.
- The approach to first find the first occurrence and then the last occurrence is sound.

Areas for Improvement:
- The `binarySearchLast` function needs to be completed to handle cases where the target is less than or greater than the mid element, similar to the `binarySearchfirst` function.
- Remove unused variables like `lowIndex` and `highIndex` to clean up the code.
- Ensure all edge cases are tested, such as when the target is not present in the array or when the array is empty.
- Consider adding more test cases to verify the correctness of the solution, especially for the `binarySearchLast` function.

PR #2309, opened 2026-02-03
3 problems completed

BharathVuppala96:


super30admin:
Strengths:
- You have understood the need for two binary searches: one for the first occurrence and one for the last.
- The structure of the code is clear, with separate functions for each binary search.

Areas for Improvement:
1. The conditions in the binary search functions for moving left or right when the target is not found at mid are incorrect. You should not use `target>=nums[l] and target<nums[m]` because it does not account for the entire sorted array. Instead, you should simply compare `nums[m]` with `target`:
   - If `nums[m] < target`, then the target must be in the right half, so set `l = m+1`.
   - If `nums[m] > target`, then the target must be in the left half, so set `h = m-1`.
   - When `nums[m] == target`, then you check for the first or last occurrence as you did.

2. In `binarySearchFirst`, when you find `nums[m] == target`, you check if it is the first occurrence by seeing if `m==0` or `nums[m-1] != target`. This is correct. Similarly for `binarySearchLast`, checking `m==len(nums)-1` or `nums[m+1] != target` is correct.

3. You should use more descriptive variable names: `low`, `high`, and `mid` instead of `l`, `h`, `m` to improve readability.

4. The condition in the `elif` branch in both binary search functions is unnecessary and causes the algorithm to fail. Remove that condition and use the standard binary search movement.

5. Also, note that in `binarySearchLast`, you should start the search from the `first` occurrence found (as in the reference solution) to optimize, but your current implementation starts from the beginning. This is not a major issue, but it can be optimized.

6. The file name should be without spaces and special characters to avoid issues in programming environments.

Corrected code for `binarySearchFirst`:
```python
def binarySearchFirst(self, nums: List[int], target: int) -> int:
    low = 0
    high = len(nums) - 1
    while low <= high:
        mid = (low + high) // 2
        if nums[mid] == target:
            if mid == 0 or nums[mid-1] != target:
                return mid
            else:
                high = mid - 1
        elif nums[mid] < target:
            low = mid + 1
        else:
            high = mid - 1
    return -1
```

Similarly for `binarySearchLast`:
```python
def binarySearchLast(self, nums: List[int], target: int) -> int:
    low = 0
    high = len(nums) - 1
    while low <= high:
        mid = (low + high) // 2
        if nums[mid] == target:
            if mid == len(nums)-1 or nums[mid+1] != target:
                return mid
            else:
                low = mid + 1
        elif nums[mid] < target:
            low = mid + 1
        else:
            high = mid - 1
    return -1
```

PR #2315, opened 2026-02-07
Complete Binary-Search-2

PrakarshKamal:


super30admin:
Your solution for the "First and Last Position" problem is excellent. It correctly implements two binary searches to find the first and last occurrence of the target. The code is efficient and meets the O(log n) requirement. Well done!

However, note that you have included solutions for other problems in the same submission. For future assignments, make sure to submit only the solution for the problem being evaluated. This helps in avoiding confusion and ensures that the evaluator focuses on the correct code.

One very minor suggestion: in the getFirst and getLast methods, you are passing the entire array and target each time. Since these are the same for both calls, you could consider making them class fields or simply not passing them (if you were to write in a class context where they are accessible). But the current approach is perfectly acceptable.

Keep up the good work!

super30admin:
Your solution for "Find the First and Last Position of an Element in Sorted Array" is correct and efficient. Well done! The code is clean and easy to understand. 

One optimization you could consider: in the `getLast` function, instead of starting the binary search from the entire array (left=0, right=n-1), you can start from the first occurrence you found. Since the array is sorted, the last occurrence must be at or after the first occurrence. So, you can call `getLast(first, n-1, nums, target)` instead. This would narrow down the search range for the last occurrence, making it slightly more efficient in practice, especially when the target appears many times. However, the worst-case time complexity remains O(log n).

Also, note that if the first occurrence is not found (i.e., `first` is -1), you can immediately return [-1, -1] without calling `getLast`. Your current code does not do this because you call both functions regardless. You should check if `first` is -1 and then return [-1, -1] without proceeding to `getLast`. This would avoid unnecessary computation.

Here's how you can modify the main method:
```java
public int[] searchRange(int[] nums, int target) {
    int n = nums.length;
    if (n == 0) return new int[] {-1, -1};
    
    int first = getFirst(0, n-1, nums, target);
    if (first == -1) {
        return new int[] {-1, -1};
    }
    int last = getLast(first, n-1, nums, target); // start from first
    return new int[] {first, last};
}
```

Additionally, your current `getFirst` and `getLast` functions might not be entirely robust if the target is not present. In `getFirst`, if the target is not found, it returns -1, which is correct. Similarly for `getLast`. However, in the `getLast` function, if we start from `first` (which is valid), it should still work correctly.

Keep up the good work!

PR #2317, opened 2026-02-08
S30 FAANMG Problem #9 | Find the First and Last Position of an Elemen…

sandy7907:
…t in given Sorted Array

### DIFF
--- a/FindMin.java
+++ b/FindMin.java
@@ -1,0 +1,25 @@
+class FindMin {
+    public int findMin(int[] nums) {
+
+        int low = 0, high = nums.length - 1, mid;
+
+        while(low <= high) {
+            mid = low + (high - low) / 2;
+
+            if(nums[low] <= nums[high]) return nums[low];
+
+            if((mid == low || nums[mid] < nums[mid - 1])
+                    && (mid == high || nums[mid] < nums[mid + 1]))
+                return nums[mid];
+
+            if(nums[mid] >= nums[low]) {
+                low = mid + 1;
+            } else {
+                high = mid - 1;
+            }
+
+        }
+
+        return -1;
+    }
+}
--- a/FindPeakElement.java
+++ b/FindPeakElement.java
@@ -1,0 +1,19 @@
+class FindPeakElement {
+    public int findPeakElement(int[] nums) {
+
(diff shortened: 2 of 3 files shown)


super30admin:
Strengths:
- The solution correctly handles edge cases, such as an empty array or when the target is not within the range of the array.
- The binary search functions are implemented correctly, with proper conditions to check for the first and last occurrences.

Areas for Improvement:
- The initial condition to check if the target is outside the array range is efficient but might be redundant because the binary search would handle it anyway. However, it is a good optimization to avoid unnecessary searches.
- In the `getFirstOccurance` function, the condition `if(mid == low || nums[mid-1] != nums[mid])` is correct, but note that when `mid == low`, accessing `nums[mid-1]` would cause an index out of bounds error. However, since the condition checks `mid == low` first (using short-circuit OR), it avoids the error. This is acceptable, but it's important to be cautious. Alternatively, you could write it as `if (mid == 0 || nums[mid-1] != target)` for clarity, similar to the reference solution.
- Similarly, in `getLastOccurance`, the condition `if(mid == high || nums[mid+1] != nums[mid])` is correct, but using `target` instead of `nums[mid]` might be more consistent, though it is equivalent since `nums[mid] == target` at that point.

Overall, the solution is excellent and meets all requirements.

super30admin:
Your solution for `SearchRange.java` is well-implemented. Here are some strengths and areas for improvement:

Strengths:
- You correctly implemented two separate binary search functions for finding the first and last occurrences, which is an efficient approach.
- The code is clean, readable, and follows good practices, such as using meaningful method names and comments (though there are no comments, the code is self-explanatory).
- You handled edge cases, such as an empty array and targets outside the array range, appropriately.

Areas for Improvement:
- While your initial check for the target being out of range is efficient, it is not strictly necessary because the binary search would return -1 anyway. However, it can improve performance in worst-case scenarios, so it's a good optimization.
- Consider adding comments to explain the logic, especially the conditions in the binary search, to make it more understandable for others.
- Ensure that you only submit the relevant code for the problem. The other files (`FindMin.java` and `FindPeakElement.java`) are not part of this problem and should be separated.

PR #2318, opened 2026-02-08
Add implementation for Binary-Search-2

rishigoswamy:


super30admin:
Your solution for leetcode34.py is well-structured and correctly implements the two binary searches to find the first and last occurrence of the target. The logic for both searches is accurate, and you've handled the boundary conditions properly (e.g., checking if mid is at the start or end of the array).

Strengths:
- You correctly used binary search to achieve O(log n) time complexity.
- The code is clean and easy to follow.
- You handled edge cases such as when the target is at the beginning or end.

Areas for improvement:
1. Remove the initial assignment of `mid` (line 95) since it is immediately reassigned in the while loop. This is redundant and might cause confusion.
2. Consider skipping the second binary search if the first search returns -1. This would make the solution slightly more efficient in the case where the target is not present.
3. Use more descriptive variable names. For example, instead of `first` and `second`, use `first_occurrence` and `last_occurrence` to make the code self-documenting.

Example improvement:
```python
class Solution:
    def searchRange(self, nums, target):
        first = -1
        low, high = 0, len(nums) - 1
        while low <= high:
            mid = (low + high) // 2
            if nums[mid] == target:
                if mid == 0 or nums[mid-1] != target:
                    first = mid
                    break
                else:
                    high = mid - 1
            elif nums[mid] > target:
                high = mid - 1
            else:
                low = mid + 1
        
        if first == -1:
            return [-1, -1]
            
        last = -1
        low, high = 0, len(nums) - 1
        while low <= high:
            mid = (low + high) // 2
            if nums[mid] == target:
                if mid == len(nums)-1 or nums[mid+1] != target:
                    last = mid
                    break
                else:
                    low = mid + 1
            elif nums[mid] > target:
                high = mid - 1
            else:
                low = mid + 1
        
        return [first, last]
```

This version avoids the second search if the target is not found and uses better variable names.

Overall, your solution is correct and efficient. The issues are minor and do not affect the correctness.

super30admin:
Your solution is on the right track and correctly implements the two binary searches. However, there are a few areas for improvement:

1. Remove the unnecessary initialization of `mid` at the top of the function. It is overwritten in the while loops.

2. After the first binary search, if `first` is still -1 (meaning the target was not found), you can immediately return [-1, -1] without running the second binary search. This will save time when the target is not present.

3. For the second binary search, you can set `low = first` instead of `low = 0` because the last occurrence must be at or after the first occurrence. This reduces the search range and makes the algorithm slightly more efficient.

4. The comments in the code are for different problems (find-minimum-in-rotated-sorted-array and find-peak-element). Please update the comments to reflect the current problem (find-first-and-last-position-of-element-in-sorted-array) to avoid confusion.

5. Consider adding a check for an empty array at the beginning to make the code even clearer, though it is not necessary because the while loops will not run when the array is empty.

Here is a slightly optimized version of your code:

```python
class Solution:
    def searchRange(self, nums, target):
        if not nums:
            return [-1, -1]
        
        first = -1
        low, high = 0, len(nums) - 1
        while low <= high:
            mid = (low + high) // 2
            if nums[mid] == target:
                if mid == 0 or nums[mid-1] != target:
                    first = mid
                    break
                else:
                    high = mid - 1
            elif nums[mid] > target:
                high = mid - 1
            else:
                low = mid + 1
        
        if first == -1:
            return [-1, -1]
        
        second = -1
        low, high = first, len(nums) - 1
        while low <= high:
            mid = (low + high) // 2
            if nums[mid] == target:
                if mid == len(nums)-1 or nums[mid+1] != target:
                    second = mid
                    break
                else:
                    low = mid + 1
            elif nums[mid] > target:
                high = mid - 1
            else:
                low = mid + 1
        
        return [first, second]
```

Overall, your solution is correct and efficient, but with these minor improvements, it can be more efficient and clearer.

PR #2319, opened 2026-02-10
Add binary search algorithms for peak finding, range searching, and m…

Balaviknesh:
…inimum in rotated sorted array

### DIFF
--- a/FindPeak.kt
+++ b/FindPeak.kt
@@ -1,0 +1,24 @@
+// Time Complexity: O(log n)
+// Space Complexity: O(1) as we are using constant space to store the indices
+class Solution {
+    fun findPeakElement(nums: IntArray): Int {
+        var low = 0
+        var high = nums.size - 1
+
+        while (low <= high) {
+            val mid = low + (high - low) / 2
+
+            if ((mid == 0 || nums[mid - 1] < nums[mid]) && (mid == nums.size - 1 || nums[mid + 1] < nums[mid])) {
+                return mid
+            }
+
+            if (nums[mid + 1] > nums[mid]) {
+                low = mid + 1
+            } else {
+                high = mid - 1
+            }
+        }
+
+        return 0
+    }
+}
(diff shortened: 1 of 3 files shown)


super30admin:
Your solution for "Find the First and Last Position of an Element in given Sorted Array" is correct and meets the required time and space complexity. Well done!

Here are some suggestions to improve the code quality:
1. Use descriptive function names: Instead of `binarySearch1` and `binarySearch2`, consider naming them `findFirstOccurrence` and `findLastOccurrence` to make the code more readable.
2. Maintain consistency in parameter order: Both binary search functions should have the same parameter order, preferably `(nums: IntArray, target: Int)`.
3. Simplify the binary search conditions: In the `else` block (when `nums[mid] != target`), you can directly check:
   - For `binarySearch1` (first occurrence): if `nums[mid] < target` then `low = mid + 1` else `high = mid - 1`. Actually, the current code does this correctly but with a condition that is a bit redundant. You can write:
   ```
   } else if (nums[mid] < target) {
       low = mid + 1
   } else {
       high = mid - 1
   }
   ```
   Similarly for `binarySearch2`.
4. Consider adding comments to explain the purpose of each function and the key conditions.

Also, note that you included solutions for other problems in the same submission. While it's great that you are practicing multiple problems, when submitting a solution for a specific problem, it's best to only include the relevant code to avoid confusion.

Keep up the good work!

super30admin:
Strengths:
- The solution correctly implements binary search for both first and last occurrences.
- The code is structured with helper functions, which is good for readability.

Areas for improvement:
1. It is inefficient to always perform both binary searches. You should check if the first occurrence is -1 and immediately return [-1, -1] without doing the second search.
2. For the last occurrence search, you can start from the first occurrence index (instead of 0) to narrow the search range. This will make the second binary search faster in practice.
3. The function names `binarySearch1` and `binarySearch2` are not descriptive. Consider renaming them to `findFirstOccurrence` and `findLastOccurrence`.
4. There is inconsistency in the parameter order of the two helper functions. Make them consistent for better code maintainability.

###

PR #2320, opened 2026-02-12
Binary Search 2 Probelms

sandy7907:


super30admin:
Your solution is well-implemented overall. The binary search functions for finding the first and last occurrences are correct and efficient. However, there is a critical issue in the initial condition check in the `searchRange` method. When the array is empty, `nums.length` is 0, so accessing `nums[0]` or `nums[nums.length-1]` will result in an `ArrayIndexOutOfBoundsException`. You should handle the empty array case separately.

To fix this, you can modify the initial check to:
```java
if (nums.length == 0) return new int[]{-1, -1};
if (target < nums[0] || target > nums[nums.length - 1]) return new int[]{-1, -1};
```

Alternatively, you can rely entirely on the binary search without the initial range check, as the binary search will naturally return -1 if the target is not found. The initial range check is an optimization that can avoid unnecessary searches, but it must be done safely.

Additionally, note that the problem constraints allow for arrays with up to 10^5 elements, so the binary search is efficient. Your implementation of the binary search is correct and follows best practices.

Another minor point: the method names `getFirstOccurance` and `getLastOccurance` have a typo ("Occurance" should be "Occurrence"). While this doesn't affect functionality, using correct spelling improves code readability and professionalism.

Overall, your solution is strong, but the edge case for an empty array needs to be addressed.

